PR #20, opened 2015-06-01
Proposal for handling shutdowns.

klauspost:
Proposal for discussion for #18 

Right now it handles:
* Reject all incoming requests once shutdown has been initiated
* Finish all incoming requests (in pre-shutdown)
* Send "shutting down" to all admins.
* Flush all outgoing messages and unsubscribe users (stage 1).

TODO: Implement flush in clients.


klauspost:
git crap. I will resubmit.

PR #53, opened 2015-12-22
Implement sharded publish queues for redis engine to improve throughput. Refs #52

banks:
As always, I'm not in love with many of the names I choose. I've opted to balance clarity and previty with shortened words but being clear about the semantics... but I'm open to suggestions.

I have this branch running now with ~5k messages a second being published and no queue backlog (with 16 shards).

Obviously we will need to document all these new features in the book once we prepare 1.3 release - @FZambia how do you normally do that? Wait until 1.3 features are final and then update it?

I also considered making the new publish more separate from the "old" api, and also considered only supporting publish/broadcast methods on it with simpler syntax but it ened up just being a lot more code complication for not much benefit.

Finally, I'm not really very pleased with the test suite for the API - I followed what was there and just push something into redis but without any assertion that it was delivered it's really not testing anything.

I may have time later to improve those tests but I needed to get something delivered for a tight deadline.

### DIFF
--- a/libcentrifugo/engineredis.go
+++ b/libcentrifugo/engineredis.go
@@ -18,13 +18,24 @@ import (
 // connected to the same Redis and load balance clients between instances.
 type RedisEngine struct {
 	sync.RWMutex
-	app           *Application
-	pool          *redis.Pool
-	psc           redis.PubSubConn
-	api           bool
-	inPubSub      bool
-	inAPI         bool
-	nPubApiShards int
+	app          *Application
+	pool         *redis.Pool
+	psc          redis.PubSubConn
+	api          bool
+	inPubSub     bool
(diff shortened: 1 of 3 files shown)


FZambia:
Thanks a lot @banks ! Will see at pull request as soon as possible.

I don't have strict strategy how to update docs, but usually I am writing and updating docs at the same day when releasing. We can create chapter in docs now in separate branch and make pull request that just merge later.

Yes, tests needs some improvements here:(

Btw what's your client code to determine queue to which you put message? Something like `crc16 mod N`?

banks:
My client code is rand.Intn(numShards) since I'm broadcasting to many channels and already have no ordering guarantees thanks to Scribe and multiple instances.

PR #410, opened 2021-01-11
JSON Web Key Support

danikarik:
## Proposed changes

In case of asymmetric JWT encryption, it's hard to update or verify incoming connection tokens due to rotation nature of keys. Current scheme of interaction produces a lot of request to refresh proxy endpoints.

According to [RFC7517](https://tools.ietf.org/html/rfc7517) there is mechanism where we can put KeyID (`kid`) into token claims and find appropriate public rsa key for verifying using key identifier. All those public keys are available by specific url. This scheme of interaction makes possible to use same jwt token as system's internal token and as Centrifugo's connection token. But of course it works only if system's jwt token fully compatible with Centrifugo's connection token and JWKS public url and token's kid are specified.

## What I suggesting

- Added new option for authentication (current `token_hmac_secret_key` and `token_rsa_public_key`) - `jwks_public_url`
- Added new jwks verifier which requests set of JWK by fetching from url above. Then using `kid` in claims it finds public key and verifies signature of token.


FZambia:
@danikarik hello, thanks - seems nice to have. I'll come with review comments very soon - need some time to get some understanding on topic.

For now I have one question - have you already tried that inside your token infrastructure?

danikarik:
Tried it in PoC. Planning to deploy it on staging within couple of days.

internal/jwtverify/token_verifier_jwt.go
```diff
@@ -58,6 +79,31 @@ type SubscribeTokenClaims struct {
...
 }
 
+type jwksClient struct{ jwks.JWKSClient }
+
+func (j *jwksClient) verify(token *jwt.Token, kid string) error {
+	if kid == "" {
+		return errEmptyKeyID
+	}
+
+	key, err := j.JWKSClient.GetEncryptionKey(kid)
```
FZambia, line 89: I think we need a key with `"use": "sig"` here - i.e. for signature verification. Looking at go-jwks code I don't see how it differentiates "sig" and "enc" - both `GetEncryptionKey ` and `GetSignatureKey` return the same.

I suppose https://auth0.com/blog/navigating-rs256-and-jwks/ has a proper example of filtering keys received from endpoint:

```javascript
      const signingKeys = keys
        .filter(key => key.use === 'sig' // JWK property `use` determines the JWK is for signature verification
                    && key.kty === 'RSA' // We are only supporting RSA (RS256)
                    && key.kid           // The `kid` must be present to be useful for later
                    && ((key.x5c && key.x5c.length) || (key.n && key.e)) // Has useful public keys
        ).map(key => {
          return { kid: key.kid, nbf: key.nbf, publicKey: certToPEM(key.x5c[0]) };
        });
```

We need to do similar things, filter by `kty` (`RSA`), also do not use JWKS at all for unsupported algorithms (I think only `RS256` is commonly used at moment).

main.go
```diff
@@ -61,6 +61,7 @@ var configDefaults = map[string]interface{}{
 	"name":                                 "",
 	"secret":                               "",
 	"token_hmac_secret_key":                "",
+	"jwks_public_url":                      "",
```
FZambia, line 64: `token_jwks_public_url` is better to be more consistent with existing option names

main.go
```diff
@@ -1060,6 +1061,11 @@ func jwtVerifierConfig() jwtverify.VerifierConfig {
 		cfg.RSAPublicKey = pubKey
 	}
 
+	jwksPublicURL := v.GetString("jwks_public_url")
+	if jwksPublicURL != "" {
+		cfg.JWKSPublicURL = jwksPublicURL
+	}
```
FZambia, line 1067: You can just replace this all with:
```go
cfg.JWKSPublicURL = v.GetString("jwks_public_url")
```

internal/jwtverify/token_verifier_jwt.go
```diff
@@ -10,34 +10,54 @@ import (
...
 
 type VerifierJWT struct {
 	mu         sync.RWMutex
+	jwksClient *jwksClient
 	algorithms *algorithms
 }
 
 var (
 	ErrTokenExpired         = errors.New("token expired")
+	ErrTokenInvalid         = errors.New("token invalid")
```
FZambia, line 58: Not used?

FZambia (commented):
Added comments, plz take a look.

main.go
FZambia: `token_jwks_public_endpoint` even better to match our proxy options that also have endpoints configurable

internal/jwtverify/token_verifier_jwt.go
```diff
@@ -46,6 +66,7 @@ type ConnectTokenClaims struct {
 	Info       json.RawMessage `json:"info,omitempty"`
 	Base64Info string          `json:"b64info,omitempty"`
 	Channels   []string        `json:"channels,omitempty"`
+	KeyID      string          `json:"kid,omitempty"`
```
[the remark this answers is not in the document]
danikarik: My mistake.

I'm preparing PR for `github.com/cristalhq/jwt/v3`, they previously had `KeyID` in `Header` in https://github.com/cristalhq/jwt/blob/v1.2.0/jwt.go#L32-L35

internal/jwtverify/token_verifier_jwt.go
```diff
@@ -10,34 +10,54 @@ import (
...
 	// HMACSecretKey is a secret key used to validate connection and subscription
 	// tokens generated using HMAC. Zero value means that HMAC tokens won't be allowed.
 	HMACSecretKey string
+
 	// RSAPublicKey is a public key used to validate connection and subscription
 	// tokens generated using RSA. Zero value means that RSA tokens won't be allowed.
 	RSAPublicKey *rsa.PublicKey
+
+	// JWKSPublicURL is a public url used to validate connection and subscription
+	// tokens generated using rotating RSA public keys. Zero value means that RSA tokens won't be allowed.
```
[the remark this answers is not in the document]
danikarik: Will be fixed.

internal/jwtverify/token_verifier_jwt.go
danikarik: Agree.

main.go
danikarik: Good.

main.go
danikarik: Will be fixed.

internal/jwtverify/token_verifier_jwt.go
danikarik: Will be fixed.

internal/jwtverify/token_verifier_jwt.go
```diff
@@ -10,34 +10,54 @@ import (
...
+
 	algorithms, err := newAlgorithms(config.HMACSecretKey, config.RSAPublicKey)
 	if err != nil {
 		panic(err)
 	}
 	verifier.algorithms = algorithms
+
+	if config.JWKSPublicURL != "" {
+		verifier.jwksClient = &jwksClient{jwks.NewDefaultClient(
+			jwks.NewWebSource(config.JWKSPublicURL),
```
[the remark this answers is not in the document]
danikarik: If write own http client which options have to be configured from config.json? Any hint?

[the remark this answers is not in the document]
FZambia: Still trying to realize what would be the best way here.

In general we don't need any extra configurations exposed except one you already added (`token_jwks_public_endpoint`). All other options can be exposed in future commits if needed.

I am trying to understand a purpose of cache in this situation. I understand that it's fine to expire some entries with time. But I don't understand why we need to periodically refresh once loaded keys - can concrete `kid` be changed? 

Even if we need to refresh: why not have much simpler cache – key is loaded once and has `ttl`, after `ttl` passed it is just removed from cache. So subsequent request will get it from jwks endpoint again. Maybe it's a good idea to load key using https://godoc.org/golang.org/x/sync/singleflight package to avoid spikes in load on jwks_endpoint. But I don't understand the purpose of 2 different durations in `s12v/go-jwks` at the moment. 

@danikarik maybe you understand more details in key update process and have some thoughts?

go.sum
```diff
@@ -68,6 +68,12 @@ github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsr
 github.com/creack/pty v1.1.7/go.mod h1:lj5s0c3V2DBrqTV7llrYr5NG6My20zk30Fl46Y7DoTY=
 github.com/cristalhq/jwt/v3 v3.0.0 h1:0PtBLOa6XEZj6powOaneZCxU3+yiFSYwVeL2wmbavnY=
 github.com/cristalhq/jwt/v3 v3.0.0/go.mod h1:XOnIXst8ozq/esy5N1XOlSyQqBd+84fxJ99FK+1jgL8=
+github.com/cristalhq/jwt/v3 v3.0.5 h1:YyWg9cAXEhtyaWBsoMyHPFg0OOEjVrwedJ4fx9b6Oqc=
```
cristaloleg, line 71: `go mod tidy` should cleanup old versions

danikarik:
Last thing is to add own JWKS wrapper.

danikarik:
Added own jwks manager to deal with cache and retrieving set from specified endpoint.

internal/jwks/cache.go
```diff
@@ -0,0 +1,25 @@
...
+	ErrCacheNotFound = errors.New("cache: value not found")
+	// ErrInvalidValue raises when type conversion to JWK has been failed.
+	ErrInvalidValue = errors.New("cache: invalid value")
+)
+
+// Cache works with cache layer.
+type Cache interface {
+	Add(ctx context.Context, key *JWK) error
+	Get(ctx context.Context, kid string) (*JWK, error)
+	Remove(ctx context.Context, kid string) error
```
FZambia, line 21: Could you please remove all unused methods from `Cache` interface?

internal/jwks/cache_ttl.go
```diff
@@ -0,0 +1,133 @@
...
+		d = time.Second
+	}
+
+	ticker := time.Tick(d)
+	go func() {
+		for {
+			select {
+			case <-ticker:
+				tc.cleanup()
+			case <-tc.stop:
```
FZambia, line 73: Is there a way to stop cache at the moment?

internal/jwks/manager.go
```diff
@@ -0,0 +1,161 @@
...
+	// If lookup is true, first try to get key from cache.
+	if m.lookup {
+		key, err := m.cache.Get(ctx, kid)
+		if err == nil {
+			return key, nil
+		}
+	}
+
+	// Otherwise fetch from public JWKS.
+	v, err, _ := m.group.Do(kid, func() (interface{}, error) {
```
FZambia, line 83: 👍

internal/jwks/manager.go
```diff
@@ -0,0 +1,161 @@
...
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/rakutentech/jwk-go/jwk"
+	"golang.org/x/sync/singleflight"
+)
+
+const (
+	_defaultRetries = 5
```
FZambia, line 17: I think 5 retries is too much for a default value, maybe 2?

internal/jwks/manager.go
```diff
@@ -0,0 +1,161 @@
...
+	_defaultTimeout = 5 * time.Second
```
FZambia, line 18: We use 1 second as default value for timeout in proxy handlers, I suggest using the same here

internal/jwtverify/token_verifier_jwt.go
```diff
@@ -58,6 +78,39 @@ type SubscribeTokenClaims struct {
...
+func (j *jwksManager) verify(token *jwt.Token) error {
+	ctx := context.Background()
+	kid := token.Header().KeyID
+
+	key, err := j.Manager.FetchKey(ctx, kid)
+	if err != nil {
+		return err
+	}
+
+	if key.Kty != "RSA" || key.Use != "sig" {
```
FZambia, line 92: I think that in scope of Centrifugo we can move Use check to manager fetch method - i.e. never deal with `enc` use at all (and never cache keys with use other than `sig`).

internal/jwks/cache_ttl.go
danikarik: https://github.com/centrifugal/centrifugo/pull/410/files#diff-f8bd2f65ee773d1c71e5c047d159b86f1729df216e4f3f30fefb59de2368eb74R136

FZambia:
Thanks @danikarik, great job 👍 Ready to be merged I suppose. Going to make some minor tweaks here before release. Also JWKS support should be documented – will ask you to review documentation as soon as I send update to it.

@cristaloleg thanks for your help!

PR #1158, opened 2026-06-14
Cache recovery for unidirectional clients with server-side subscriptions

FZambia:
Adds a namespace channel option **`auto_cache_recover`** that makes Centrifugo automatically
initiate cache recovery for all subscriptions in the namespace on subscribe, without requiring
the subscriber to request it. In cache recovery mode this delivers the latest publication on
every (re)subscribe — so client-side subscribers no longer need to provide an empty `since`, and
it also works for **server-side subscriptions of unidirectional clients** (SSE, HTTP-streaming,
unidirectional WebSocket) that can't request recovery themselves and may not even know channel
names.

It also adds a per-subscription **`cache_recover`** field to the connect proxy result, so a
connect proxy can enable the same behavior for specific server-side subscriptions instead of
namespace-wide.

## Motivation

Cache recovery mode lets a channel behave like a real-time key-value store: the latest
publication is delivered as the first event on subscribe. Until now this required the client to
opt in by sending an empty `since`. That doesn't work for:

- server-side subscriptions (subscriptions decided by a connect proxy `subs`/`channels` or by a
  JWT), and
- unidirectional clients, which only send a token and may not know channel names.

`auto_cache_recover` covers these cases by letting the server force cache recovery for the
namespace (or, via the proxy, per subscription).

## Changes

- New namespace option `auto_cache_recover` (`configtypes.ChannelOptions.AutoCacheRecover`).
  Validation requires `force_recovery` and `force_recovery_mode: cache`; otherwise config
  validation fails fast.
- Client handler sets the forced-recovery flag on server-side subscriptions (from a connect
  proxy or JWT) and on client subscriptions in the namespace, gated to cache recovery mode.
- Connect proxy: new `cache_recover` field on the per-subscription `SubscribeOptions` message
  (`proxy.proto`, regenerated `proxy.pb.go`). When set, Centrifugo forces cache recovery for
  that server-side subscription — the server-side equivalent of an empty `since`.

### Configuration example

```json title="config.json"
{
  "channel": {
    "namespaces": [
      {
        "name": "example",
        "force_recovery": true,
        "force_recovery_mode": "cache",
        "auto_cache_recover": true,
        "history_size": 1,
        "history_ttl": "1h"
      }
    ]
  }
}
```

A unidirectional listener whose channels are decided server-side then receives the latest
publication per channel on every reconnect, without knowing channel names or stream positions.

## Dependency

Requires the matching Centrifuge change that renames `SubscribeOptions.Recover` →
`AutoCacheRecover` and scopes server-forced recovery to cache recovery mode (ignored in stream
mode). Centrifugo only enables the flag in cache mode, so the two are consistent: the option is
cache-only on both layers.

## Behavior

- Cache recovery mode: with `auto_cache_recover` enabled, every (re)subscribe delivers the
  latest publication (`recovered: true` when a publication exists, `false` on an empty/expired
  cache). Applies to client-side and server-side subscriptions in the namespace.
- The option requires `force_recovery` + `force_recovery_mode: cache`; any other combination is
  rejected at config validation time. It has no effect outside cache recovery mode.
- No behavior change for namespaces that don't set the option.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sqs v1.43.2
 	github.com/aws/aws-sdk-go-v2/service/sts v1.43.2
 	github.com/aws/smithy-go v1.27.2
-	github.com/centrifugal/centrifuge v0.38.1-0.20260614052230-24996932b8ed
+	github.com/centrifugal/centrifuge v0.38.1-0.20260615180132-f369d8fe1105
 	github.com/centrifugal/protocol v0.19.2
 	github.com/cristalhq/jwt/v5 v5.4.0
 	github.com/go-viper/mapstructure/v2 v2.5.0
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/service/sqs v1.43.2
 	github.com/aws/aws-sdk-go-v2/service/sts v1.43.2
 	github.com/aws/smithy-go v1.27.2
-	github.com/centrifugal/centrifuge v0.38.1-0.20260531113025-945c9cce05ad
+	github.com/centrifugal/centrifuge v0.38.1-0.20260614052230-24996932b8ed
 	github.com/centrifugal/protocol v0.19.2
 	github.com/cristalhq/jwt/v5 v5.4.0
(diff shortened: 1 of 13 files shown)


FZambia:
Updated PR description with to reflect changes in 707b37464c6e757a95359648428739bbe533e10a

PR #1166, opened 2026-07-03
Fix/nats jetstream recreate on consumer deleted

thuy-le-kafi:
Follow-up to our Telegram discussion.

  Problem

  The NATS JetStream async consumer only re-creates itself on ErrNoHeartbeat. When the durable consumer (or its whole stream) is deleted on the server side, pull requests keep failing with nats: consumer deleted, which is only logged:

```
{"level":"error","consumer":"nats_market_quote","error":"nats: consumer deleted","message":"error during consuming"}
```

  Centrifugo then never consumes from that stream again until the process is restarted, while looking perfectly healthy. ErrNoHeartbeat is never triggered in this case: the failure surfaces as ErrConsumerDeleted instead, which was previously only logged and not acted upon.

  This happens in practice when a stream has to be deleted and re-created e.g., to change immutable fields (subjects, storage, limits) or when an in-memory stream is lost after a NATS restart. The stream comes back quickly, but the consumer doesn't, and the only recovery is a manual restart of all Centrifugo nodes.

  Reproducing steps

  1. Run Centrifugo with a nats_jetstream consumer attached to a stream.
  2. nats stream rm <STREAM> - Centrifugo starts logging error during consuming ... nats: consumer deleted forever.
  3. nats stream add <STREAM> (same config) - stream is back, but Centrifugo never consumes again until restarted. Can check by: nats consumer report <STREAM>.

  Fix

  ErrConsumerDeleted, ErrConsumerNotFound and ErrStreamNotFound now also trigger the existing recreation loop (same path as heartbeat loss). The loop already retries with backoff, so while the stream is absent it just keeps retrying; once the stream is back, the durable is re-created with the configured deliver_policy and consuming resumes - same semantics as a process restart, just automatic.

  Added an integration test covering the flow: consume -> delete stream -> re-create stream -> messages flow again without restart. The test is under the existing //go:build integration tag (same as the current NATS test) and passes against a local nats-server -js.

### DIFF
--- a/internal/consuming/nats_jetstream.go
+++ b/internal/consuming/nats_jetstream.go
@@ -35,7 +35,8 @@ type NatsJetStreamConsumer struct {
 
 // NewNatsJetStreamConsumer creates a new NatsJetStreamConsumer instance.
 // On startup if creating a consumer fails then an error is returned.
-// Later at runtime, if a heartbeat error occurs the consumer is re-created endlessly.
+// Later at runtime, if a heartbeat error occurs or the consumer/stream is
+// deleted on the server side the consumer is re-created endlessly.
 func NewNatsJetStreamConsumer(
 	cfg NatsJetStreamConsumerConfig,
 	dispatcher Dispatcher,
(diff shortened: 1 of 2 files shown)


FZambia:
Thanks @thuy-le-kafi !

FZambia (approved):

PR #1186, opened 2026-07-10
go 1.26.5, ups deps, drop unused option

FZambia:
Subj


FZambia:
Forgot CI files - fixed in 6568dad127e53e5a3de1dc4b0737e8bc50f2bccd

PR #1207, opened 2026-08-18
Fix client.connection_rate_limit refilling at 1 req/sec

### LINKED ISSUE #1206: [bug] client.connection_rate_limit refills at 1 request/second regardless of configured value
## Describe the bug

`client.connection_rate_limit` is documented as the maximum number of HTTP requests to establish a new real-time connection that a node accepts **per second**.

The limiter is currently constructed as:

```go
rate.NewLimiter(rate.Every(time.Second), connRateLimit)
```

In `golang.org/x/time/rate`, the first argument is the sustained rate in events per second. `rate.Every(time.Second)` therefore produces a sustained rate of `1`, while `connRateLimit` is used only as the burst size.

As a result, configuring `client.connection_rate_limit` to `N` allows an initial burst of `N`, but after that burst is depleted the limiter refills at only one connection request per second. Requests outside that effective rate receive HTTP 503.

Documentation: https://centrifugal.dev/docs/server/configuration#clientconnection_rate_limit

## Versions

- Observed with Centrifugo v6.8.4.
- Confirmed the same implementation is present in v6.9.2 and current `master` as of 2026-08-18.
- Operating system: Linux arm64 container.
- Client library: not applicable; this is server HTTP middleware behavior.

v6.9.2 source: https://github.com/centrifugal/centrifugo/blob/v6.9.2/internal/middleware/connlimit.go

## Steps to reproduce

Add this focused test to the `internal/middleware` package:

```go
func TestConnectionRateLimiterUsesConfiguredPerSecondRate(t *testing.T) {
    const configuredLimit = 100

    limiter := connectionRateLimiter(configuredLimit)
    if got, want := limiter.Limit(), rate.Limit(configuredLimit); got != want {
        t.Fatalf("sustained rate = %v, want %v", got, want)
    }
    if got, want := limiter.Burst(), configuredLimit; got != want {
        t.Fatalf("burst = %d, want %d", got, want)
    }
}
```

The test fails with:

```text
sustained rate = 1, want 100
```

The burst assertion passes with `100`.

## Expected behavior

A configured value of `100` should create a limiter with a sustained rate of 100 requests per second and a burst size of 100.

## Suggested fix

```go
rate.NewLimiter(rate.Limit(connRateLimit), connRateLimit)
```

This preserves the existing burst size while making the sustained rate match the documented configuration value. A regression test can assert both `Limiter.Limit()` and `Limiter.Burst()`.

moses-mbaga:
## Proposed changes

- Use `client.connection_rate_limit` as the token bucket's sustained per-second rate.
- Preserve the configured value as the burst size.
- Extend the existing connection-limit middleware test to assert both values.

The limiter currently uses `rate.Every(time.Second)`, which always refills at one request per second regardless of the configured limit. Converting the configured integer to `rate.Limit` makes the sustained rate match the documented behavior.

## Validation

- `go test -race ./internal/middleware -count=1`
- `go build ./...`
- `golangci-lint run --timeout 10m0s` with the repository-pinned v2.11.1

Fixes #1206

### DIFF
--- a/internal/middleware/connlimit.go
+++ b/internal/middleware/connlimit.go
@@ -54,7 +54,7 @@ func (l *ConnLimit) Middleware(h http.Handler) http.Handler {
 
 func connectionRateLimiter(connRateLimit int) *rate.Limiter {
 	if connRateLimit > 0 {
-		return rate.NewLimiter(rate.Every(time.Second), connRateLimit)
+		return rate.NewLimiter(rate.Limit(connRateLimit), connRateLimit)
 	}
 	return nil
 }
--- a/internal/middleware/connlimit_test.go
+++ b/internal/middleware/connlimit_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/centrifugal/centrifugo/v6/internal/tools"
 
 	"github.com/stretchr/testify/require"
+	"golang.org/x/time/rate"
 )
 
 func TestConnLimit_ConnectionRate(t *testing.T) {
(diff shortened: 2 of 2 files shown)


FZambia:
Hi @moses-mbaga 

Many thanks for the report and the fix!

